PR #4, opened 2015-01-15
passwords.py file

bheesham:
Changed to using a `passwords.py` file for storing credentials. This should take care of the whole "passwords being public" thing.

[See this issue.](https://github.com/rust-lang/rust/issues/20629)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ master/gitpoller*
 master/http.log
 master/master.cfg.txt
 master/master.cfg.sample
+master/passwords.py
+master/passwords.pyc
 master/public_html
 master/runtests
 master/slave-list.txt
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -1672,22 +1672,31 @@ c['status'] = []
 from buildbot.status import html
 from buildbot.status.web import authz, auth
 
-# we add one trivial user to prevent spambots from hitting the web UI
-# (yes, this has happened)
-users = [("rust", "rust")]
+# Import users, one that is able to build anything (any-build),
+# and another that is not allowed to build dists (no-dist).
+import passwords
+reload(passwords)
+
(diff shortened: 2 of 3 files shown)


bheesham:
Uuuhhh, ignore this for now. There shouldn't be a merge conflict.

nikomatsakis:
I know very little about this. Changing reviewer to @brson

brson:
I'm not sure this is sufficient. It will allow us to keep the passwords hidden from the source code, but it leaves a few problems:

* It doesn't create any distinction between what credentials are required for triggering e.g. try builds vs those for dist builds, and dist builds are the only ones that need to be restricted to just a few people.
* Passwords are still transmitted over plain text to buildbot.

brson:
A concrete change here might be to distinguish credentials for dist builds from all others. I think we should probably ask the people in #buildbot on freenode about best practices, though perhaps we just need to bite the bullet and put buildbot behind an https proxy.

bheesham:
Putting buildbot behind nginx or something similar was what I was going to
suggest to get around the whole "transferral of passwords over plaintext".

I'll ask on #buildbot later today or tomorrow if I don't get the
information I want from the Buildbot wiki.

bheesham:
It looks like there's some fine-grain control available for this: http://docs.buildbot.net/current/manual/cfg-statustargets.html#authentication

I'll test out the `HTPasswdAuth` method, because that would probably be easier to manage.

brson:
passwords.py seems good

bheesham:
So currently the two changes are:
  * Passwords are in a `passwords.py` file.
  * There are two different users, one that is not allowed to touch anything that has "dist" in it's name, and another one that is.

brson:
Looks good, thanks. Testing it now.

brson:
Thanks @bheesham!

PR #9, opened 2015-02-23
Kill _all_ builders when a build fails

dotdash:
Currently builders that shouldn't cause the others to be killed aren't
killed themselves either when one of the gated builders fails. That
doesn't seem right.


brson:
Rebased, merged and deployed. Thanks!

PR #13, opened 2015-05-15
Add initial MSVC bot configuration

alexcrichton:
* Rename all current windows builders to `win-gnu-XX`
* Currently bootstrap MSVC with a two-host build (other host is `windows-gnu`)
* Don't run tests for MSVC just yet
* Don't gate on MSVC just yet


brson:
r=me

PR #14, opened 2015-05-19
adding config for cargo-bitrig-64

dhuseby:
I think this correctly adds the configuration for cargo-bitrig-64.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -143,7 +143,7 @@ auto_platforms_prod = [
 try_platforms_prod = ["linux", "win-gnu-32", "win-gnu-64", "bsd", "mac"]
 snap_platforms_prod = ["linux", "win-gnu-32", "win-gnu-64", "bsd", "mac", "bitrig-64"]
 dist_platforms_prod = ["linux", "win-gnu-32", "win-gnu-64", "mac"]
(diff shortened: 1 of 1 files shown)


brson:
Thanks! I'll deploy this tomorrow.

master/master.cfg
```diff
@@ -191,6 +191,7 @@ nogate_builders = [
     "auto-bitrig-64-opt", 
     "auto-linux-musl-64-opt",
     "auto-win-msvc-64-opt",
+    "cargo-bitrig-64"
```
alexcrichton: I think that this array is only used for the auto builders themselves, not necessarily the cargo builders unfortunately. I also unfortunately am not familiar with how `nogate_builders` works :(

dhuseby:
Hrm...I looked more closely and it looks like we only gate on builders with names that start with "auto": https://github.com/rust-lang/rust-buildbot/blob/master/master/master.cfg#L1824

The BSKiller class takes a lambda filter that only considers builds with names that start with "auto" and are not in the nogate_builders list.

So my change to the nogate_builders list doesn't do anything.

dhuseby:
I just removed the change to nogate_builders since it doesn't do what I think it does.

dhuseby:
@brson status?

brson:
I'm going to deploy this with some other changes tonight.

brson:
Deployed.

PR #16, opened 2015-05-27
Request spot instances

seankelly:
The max price is set to $0.05. In us-west-1, one region consistently has a price of around $0.032 for m3.xlarge.

I've never used EC2 instances but this appears to be all that is necessary to request spot instances.


brson:
Wow, thanks for this patch! This is definitely something we want to pursue.

Unfortunately, I've tried exactly this before and there's more to it. The first error we're going to see is that the AWS account buildbot uses is going to need to have new permissions added. The one time I tried this I could not get it to work correctly.

It's going to require a lot of validation in the dev environment too to have confidence that spot instances will work reliably.

cc @edunham

brson:
Closing since this is old and will take a lot of consideration to pull off.

PR #26, opened 2015-09-06
environment adaptation for openbsd

semarie:
Use gcc-4.9 (from ports), instead of gcc-4.2 (from system base), for
building.

It is done by using egcc/eg++ as compiler and linker. The package
lang/gcc/4.9 should be installed. The binaries are in /usr/local/bin
(normally included in PATH). Set CC, CXX and RUSTFLAGS for pass `-C
linker' option to rust compiler.

cc: @dhuseby @alexcrichton


alexcrichton:
Is this something that should instead go into the standard build system? E.g. if this is required on basically all OpenBSD builders it may make more sense to check into the tree rather than buildbot here.

semarie:
gcc-4.9 is a third-party software. detecting it in configure may be doable. but does it possible to set `RUSTFLAGS` from configure ?

brson:
If I understand correctly, `egcc` etc. are not real names for these compilers, but are expected by rust-buildbot to be symlinked to the proper ports build.

There are other places in this file where we instead from `PATH` for specific builders so that the proper gcc is invoked, accomplishing this same thing, but with a different strategy. Now I actually like *this* strategy of builder-specific gcc names pretty well, but having two strategies is kind of bad. Alternately, we could do nothing here to teach rust-buildbot about the location of the correct compiler and force the slave to setup the `PATH` variable correctly.

Feeling wishy-washy here.

alexcrichton:
I'd be fine munging `PATH` but something like `-C linker=egcc` I think needs to be part of our build system instead of set here instead (or it's set on the slave)

dhuseby:
Why don't we just patch the configure script to do the right thing for OpenBSD?  That's what I did for Bitrig and turning off jemalloc.

dhuseby:
BTW, I just removed the "bsd" clause just about the lines added in this patch.  I think the correct way to detect and use the correct compiler is through the configure script.

dhuseby:
I vote :-1:

dhuseby:
from @semarie readme: `warning: check the value of your data limit (ulimit -d): building rustc will need 1.5G.`  a test for this also needs to be added to the configure script.

dhuseby:
freebsd also comes with the corefiles maxsize set to infinite which is bad for the test pass.  i need to add a check in configure for that as well.

semarie:
@brson `egcc` isn't a symlink under openbsd. gcc-4.9 is installed as third-party, with a binary named as it (under /usr/local/bin), in order to not conflict with gcc from base (which is almost a fork of gcc-4.2), and which resid under /usr/bin. The fact OpenBSD don't have recent gcc in base is partially due to licence problem (GPL3).

The same problem is encountered for libstdc++. It is the stdc++ from base. gcc-4.9 install libestdc++. And as rust is compiled without default libraries, but add it itself, currently the name of stdc++ is wrong (I still need patching `src/etc/mklldeps.py` to replace stdc++ by estdc++).

For compiler name, the play with `PATH` could be possible. It is the way that third-party programs are build under OpenBSD (aka ports). The port framework make a `bin` directory and make symbolic link pointing to the right compiler (clang, gcc-4.9, gcc-4.8, gcc from base, depending the build configuration of the port), and add this directory in front of `PATH`.

But from rust perpective, it still requires the user to do something particular for building.

@dhuseby I am ok to look at configure to do some sort of detection, but I don't want to make the configure work under *my* (and current) configuration only. I have some plans for clang under OpenBSD, so just adding "if openbsd, then CC=egcc" would be bad. Flexibility is really important here.

dhuseby:
@semarie The OpenBSD buildbot is a stock/unmodified OpenBSD install.  I have just installed the bare minimum to build Rust from source (e.g. gcc-4.9, etc).  The general algorithm that configure uses to detect the compiler is like this:

```
if (--enable-clang == true) {
  if (clang found) {
    configure to use clang
  } else {
    fail
  }
} else {
  if (gcc found) {
    configure to use gcc
  } else if (clang found) {
    configure to use clang
  } else {
    fail
  }
}
```

So on OS's with only clang (e.g. Bitrig), there is a block that says, "if Bitrig, force --enable-clang to TRUE".  For OpenBSD, you should add a block in the gcc detection code that says, "If OpenBSD, only look for 'egcc'".

That way, if you configure *without* --enable-clang, the configure script will look for egcc, if it doesn't find egcc, it will search for clang.  If you configure *with* --enable-clang, it will only search for clang.

A change like that should meet the requirements for regular users, your port, and the buildbot.

brson:
@semarie Thanks for the clarification.

Since egcc is the standard name for this compiler, I agree it would be great for the Rust build to know about it explicitly on OpenBSD, and @dhuseby's [solution] sounds right. @alexcrichton Does this sound right to you? @semarie Do you mind preparing a patch against Rust's configure script that detects egcc on OpenBSD?

alexcrichton:
Sounds good to me!

semarie:
ok, I will go in this direction. thanks.

PR #27, opened 2015-09-08
unnecessary force of clang++ for bsd platforms

dhuseby:
it is unnecessary to for clang++ on bsd platforms.  the config script should correctly detect CC and CXX on Bitrig, FreeBSD, DragonflyBSD, OpenBSD, and NetBSD.


dhuseby:
@alexcrichton this will get the dragonflybsd buildbot green.

PR #39, opened 2015-10-20
Move all linux slaves to docker images

alexcrichton:
This updates all our linux slaves to be driven from docker images (defined in this repo). Configuration for the master was tweaked here and there for new paths and such, but otherwise all changes here are just creating the docker images. All-in-all, we've now got 4 docker images:

* android - this is used for the `auto-linux-64-x-android-t` and `nightly-dist-rustc-arm-android` builders. This may be able to get merged with the linux image below in the future, but it would require figuring out how to isolate emulators. Right now two builders use this image and it only works because the slaves are limited to just one build.
* dist - our CentOS 5 image with a "very old libc". This has tons of custom-built software, but it's used to produce cargo and rustc nightlies. This is not used for MUSL or cross nightlies.
* linux - the standard Ubuntu 14.04 image we run all linux tests in. This is also used to build MUSL as it's installed on the side.
* linux-cross - a 14.04 image to build all the cross-std architectures and such. This is not included in the normal linux image because Ubuntu's current situation means that you can't have a multilib (32 + 64 bit) gcc while also having all the cross compilers. Instead we just have a 64-bit gcc with all the cross compilers.


alexcrichton:
May not want to merge just yet as the final tests are still running, but wanted to get some initial feedback

edunham:
:+1: if the tests behave as expected

brson:
r+

alexcrichton:
Everything passed in dev!

PR #40, opened 2015-10-23
Try adding travis config

alexcrichton:


alexcrichton:
cc @edunham as well

For future reference I had to disable travis on PRs because we need our AWS credentials for `checkconfig` to work, this means we'll just have to be notified after-the-fact of a push if something broke.

Issue #45, opened 2015-11-17
Docker slaves need Valgrind


edunham:

nrc:
cc @nrc

nrc:
This is good to go now that rust-lang/#29896 is fixed

PR #46, opened 2015-11-23
make Docker install valgrind

edunham:
Checked for a `valgrind --version` on prod-slave-linux-3 but it's not
installed there.

fixes https://github.com/rust-lang/rust-buildbot/issues/45


alexcrichton:
We'll want to ensure this works ok in dev on 32 and 64 bit as I think hokiness was required in the past to get a 32-bit valgrind working. I've pushed this with the tag 2015-11-23 as well

PR #47, opened 2015-11-25
Fix bitrot in dist/Dockerfile

rillian:
This didn't build for me. To get it working again:

- Update cmake source url.
- Move curl build to top so we can use SNI for cmake source url.
- Actually build curl against the openssl we build.

### DIFF
--- a/slaves/dist/Dockerfile
+++ b/slaves/dist/Dockerfile
@@ -40,19 +40,19 @@ RUN sh build_openssl.sh && rm -rf /build
 COPY dist/build_curl.sh /build/
 RUN sh build_curl.sh
 
-# binutils < 2.22 has a bug where the 32-bit executables it generates
-# immediately segfault in Rust, so we need to install our own binutils.
-#
-# See https://github.com/rust-lang/rust/issues/20440 for more info
-COPY dist/build_binutils.sh /build/
-RUN sh build_binutils.sh && rm -rf /build
-
 # Install gcc 4.7 which has C++11 support which is required by LLVM
 #
 # After we're done building we erase the binutils/gcc installs from CentOS to
 # ensure that we always use the ones that we just built.
 COPY dist/build_gcc.sh /build/
 RUN sh build_gcc.sh && rm -rf /build
+
+# binutils < 2.22 has a bug where the 32-bit executables it generates
+# immediately segfault in Rust, so we need to install our own binutils.
+#
+# See https://github.com/rust-lang/rust/issues/20440 for more info
+COPY dist/build_binutils.sh /build/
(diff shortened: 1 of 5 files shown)


edunham:
r+ 
Thanks for the fixes!

alexcrichton:
Could as much as possible be placed after building gcc? When tweaking these scripts it's always unfortunate to have to rebuild gcc a bunch of times.

rillian:
If binutils is only needed for rust, I could move gcc one step earlier, after just openssl and curl.

I'm loath to move it more than that. The gcc build doesn't currently require SNI, so we could use the system curl, but centos 5 has openssl 0.9.8e (hopefully patched!!) which is end-of-life in 5 weeks. Given Centos 5 is only receiving critical fixes at this point, I worry about the weakness this places on our verification steps.

Would it help to rebuild openssl and curl with the new gcc? I don't actually understand what would trigger the need to rebuild gcc with the current order.

alexcrichton:
I'm not really concerned about the security of images like this, it's built quite rarely so the chance of something trying to exploit an OpenSSL bug or something like that seems *super* unlikely. I'd mostly just favor tweaks to the image to make development easier, and building gcc is a pain

rillian:
These scripts build the compiler which builds the rust toolchain. An attacker could use any vulnerability here to trojan all software built with rust. Even if a targeted attack is unlikely, the consequences are very serious. An attacker can choose the easiest path; we have the responsibility of shoring up all of them.

In any case, this seems like a separate issue from fixing the bitrot. I propose we merge this based on @edunham's r+. I'll finish testing the binutils move and post it as a separate PR. We can continue discussing it there. (Believe me, I know rebuilding gcc is a pain; I've been doing it all week!) 

Is your concern that openssl revs more frequently  than gcc, and updating it means the gcc build will no longer be in docker's cache?

rillian:
As far as I can tell, reordering binutils and gcc works. r? @alexcrichton, @edunham.

alexcrichton:
Er yes sorry, to be clear I'm worried about build times not security right here. If we actually want to be "super duper secure" then we shouldn't download anything and just bundle it all in this repo. I suspect openssl will change quite a bit as they release new versions to fix older bugs, and I suspect that gcc will change very rarely as we just want the bare minimum that works.

rillian:
That sounds like an r-. I'll flip the order and retest, probably tomorrow when I'm back at my desktop.

alexcrichton:
Ok, thanks @rillian!

rillian:
Moving gcc to the first step worked for me. r? alexcrichton.

alexcrichton:
Thanks!

PR #48, opened 2015-12-03
Verify checksums after downloading source.

rillian:
Keep a 256 bit SHA-2 checksum for the source packages we build
in each build script, and verify it after download. This protects
against corrupt archives and gives us another way to track how
the build environment is created.

Using the `tee >(sha256sum)` trick to pass the curl output to a
subshell as well as tar requires bash. To do this in portable
shell we need to write the tarball to disk, increasing temporary
image footprint.


rillian:
This is intended to alleviate my security concerns about building gcc before openssl and curl.
r? @edunham. cc @alexcrichton

edunham:
LGTM. @alexcrichton , do you know of any edge cases in the buildslaves where `sh` doesn't already mean `bash`?

alexcrichton:
Looks good to me, thanks!

PR #54, opened 2016-01-23
Set up iOS slaves for nightly builds

alexcrichton:
Few changes here:

* There are new auto builders for the iOS builds
* Mac slaves are enabled in dev
* Nightly triggers are disabled in all but prod 
* Dist builders are set up for iOS (building standard libraries)


alexcrichton:
I ran this through a build in dev and miraculously they worked the first time....

* http://54.176.156.253/builders/auto-mac-ios-opt/builds/0
* http://54.176.156.253/builders/nightly-dist-rustc-mac-ios/builds/0

PR #57, opened 2016-01-29
More linux-cross targets

alexcrichton:
* powerpc-unknown-linux-gnu
* powerpc64le-unknown-linux-gnu
* x86_64-rumprun-netbsd


alexcrichton:
Some refactoring here as well.

rumprun is still disabled while the libstd build comes back online, but powerpc targets are added.

PR #59, opened 2016-02-10
fix curl download

rillian:
Build script was bit-rotted by haxx.se adding an http->https redirect, which the centos 5 curl is too old to follow. @bagder was kind enough to add an http-only alternative url.

Update to the latest version while we're here.


alexcrichton:
Thanks!

PR #61, opened 2016-02-12
Add auto builders for rustbuild build system

alexcrichton:
Don't gate on them just yet as we need to make sure they actually work.


alexcrichton:
Only covers half the tier 1 platforms (no 32/64 for each platform  yet), but should cover the basics of Linux, OSX, MinGW, and MSVC

PR #62, opened 2016-02-23
Write cargo date file for use by v2 manifests

edunham:
This change has been working in dev.


alexcrichton:
lgtm, @brson think we should do this for other components as well? I figure the manifest takes care of it for external tooling though and this is just for our own internal usage, so I'd be fine having this be cargo-specific

PR #65, opened 2016-03-07
Add support for Android NDK 16

kanerogers:
This PR adds support for Android NDK 16, as discussed in https://github.com/rust-lang/rust/issues/31119#issuecomment-192957651 :smile:

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -817,7 +817,7 @@ class ConfigCommand(object):
 
         # Set up the path to the android NDK
         if testing_android:
-            s += " --android-cross-path=/android/ndk-arm-18"
+            s += " --android-cross-path=/android/ndk-arm-16"
             s += " --disable-docs"
 
         if "release-channel" in props:
--- a/slaves/android/install-ndk.sh
+++ b/slaves/android/install-ndk.sh
@@ -8,6 +8,13 @@ set -ex
 curl -O http://dl.google.com/android/ndk/android-ndk-r10e-linux-x86_64.bin
 chmod +x android-ndk-r10e-linux-x86_64.bin
 ./android-ndk-r10e-linux-x86_64.bin > /dev/null
+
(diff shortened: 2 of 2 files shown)


alexcrichton:
Perhaps this could be tested out first? I think that this'll also need to install an API level 16 emulator and also switch to using that emulator instead of the arm-18 emulator.

kanerogers:
No problem Alex! Just point me in the right direction and I'll get it done.

alexcrichton:
Ah yeah sure. What you'll basically want to do is emulate what the bots themselves do. [Our buildbot](http://buildbot.rust-lang.org/) has a ton of builders, but you're most likely only interested in the [android one](http://buildbot.rust-lang.org/builders/auto-linux-64-x-android-t). You can take a look at a [recent successful build](http://buildbot.rust-lang.org/builders/auto-linux-64-x-android-t/builds/8272) to get an idea of how we run the build.

Basically what you'll want to do is:

1. Modify the slave images here to install a level 16 NDK
2. Modify the image to have a level 16 emulator (may need to install some more things, see `install-sdk.sh`
3. Spin up the image
4. Clone Rust
5. Configure rust for cross compilation
6. Run the build
7. Start up right emulator in the background
8. Run the tests in the emulator

The builder should show you the commands that are needed to run all this, but once that's all up and running this should be good to go! We'll likely want to land the standard library changes first, and then I'll merge these changes here, test them in dev, merge them to master, and deploy them to prod.

kanerogers:
Got it. Leave it with me. :- )

Thanks so much for all the help!

brson:
Neat!

rillian:
I presume #99 would resolve this too.

alexcrichton:
I'm hoping to add API level 9 support soon, so hopefully that'll be more than enough to target API level 16, so I'm gonna close this for now. Thanks though for the PR @kanerogers!

kanerogers:
No worries, Alex! Sorry I didn't get a chance to take this any further. :- (

PR #66, opened 2016-03-07
Build v2 manifests

brson:
This builds v2 manifests during `finish_dists` for the "rust" component. It adds the file `build-rust-manifest.py`. It's well-commented so read it for details.

This adds a new `channel-$component-$channel-date.txt` file during `finish_dist` as part of v1 manifest generation. Then, when building the "rust" component, `build-rust-manifest` discovers the cargo and rustc archive dates from that file on s3. It then validates the existence of the cargo and rustc components by pinging s3, and the final rust components by looking at the local disk.

Two ugly points to note:

* `build-rust-manifest.py` is invoked [by absolute path](https://github.com/brson/rust-buildbot/blob/wip/master/master.cfg#L1387).
* [There's something of a security hole](https://github.com/brson/rust-buildbot/blob/wip/master/build-rust-manifest.py#L23) in the plaintext s3 validation. I think this is not in practice something we need to be too concerned with now.

r? @alexcrichton cc @edunham 

I'd like to deploy this tonight.


brson:
[Example](http://dev-static-rust-lang-org.s3-website-us-west-1.amazonaws.com/dist/channel-rust-nightly.toml).

master/master.cfg
```diff
@@ -1364,19 +1370,40 @@ def finish_dist(f, local_dist_dir, dist_subdirs, s3_addy, remote_dist_dir, compo
...
-                                     command=["sh", "-c", cargo_date_cmd] ))
+    # Build the channel manifest v2, which is a toml file describing the components
+    # of Rust. This only works for the 'rust' component, which contains various
+    # artifacts of other builds. It depends on the v1 manifests and channel build
+    # date of the 'cargo' and 'rustc' channels.
+    if component == "rust":
+        manifest_v2_name = "channel-" + component + "-" + channel + ".toml"
+        manifest_v2_file = final_dist_dir + "/" + manifest_v2_name
+        # FIXME: This is a poor way to find the build-rust-manifest script
+        manifest_v2_cmd = ("python $HOME/rust-buildbot/master/build-rust-manifest.py " +
```
alexcrichton: I'm not 100% privy to how python works, but couldn't this be `import build_rust_manifest` at the top of this file? That can import the local module directly, right? That may also help in marshalling arguments back and forth

brson: It's possible. The reasons I didn't are:

* Assigning the arguments to globals in the python script is nice. I did spend a few minutes threading them around and decided I didn't want to.
* I was hopeful that running this as a MasterShellCommand would automatically do it asynchronously on another thread. Blocking here would be bad.

master/build-rust-manifest.py
```diff
@@ -0,0 +1,489 @@
...
+    "i686-pc-windows-msvc",
+    "x86_64-apple-darwin",
+    "i686-apple-darwin",
+])
+
+# These are the platforms we produce standard libraries for
+target_list = sorted([
+    "aarch64-unknown-linux-gnu",
+    "arm-linux-androideabi",
+    "arm-unknown-linux-gnueabif",
```
alexcrichton: "gnueabi"

master/build-rust-manifest.py
```diff
@@ -0,0 +1,489 @@
...
+    "arm-unknown-linux-gnueabihf",
+    "i686-apple-darwin",
+    "i686-pc-windows-gnu",
+    "i686-pc-windows-msvc",
+    "i686-unknown-linux-gnu",
+    "mips-unknown-linux",
```
alexcrichton: needs "-gnu" at the end (same with mipsel below)

master/master.cfg
edunham: Using globals will be unpleasant if we ever attempt to unit test the script. Based on what I've learned from attempting to do the same thing earlier, I think having the script separate may make it easier to add debugging information and tests in the future

master/build-rust-manifest.py
```diff
@@ -0,0 +1,489 @@
...
+    "mipsel-unknown-linux",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-gnu",
+    "x86_64-pc-windows-msvc",
+    "x86_64-unknown-linux-gnu",
+    "x86_64-unknown-linux-musl",
+])
```
alexcrichton: * aarch64-apple-ios
* armv7-apple-ios
* armv7-unknown-linux-gnueabihf
* armv7s-apple-ios
* i386-apple-ios
* powerpc-unknown-linux-gnu
* powerpc64-unknown-linux-gnu
* powerpc64le-unknown-linux-gnu
* x86_64-apple-ios
* x86_64-rumprun-netbsd

alexcrichton: er, meant to prefix that with "Some other targets":

alexcrichton: There's also a list of these kinda in master.cfg which we'll need to keep in sync, but probably too difficult for now to refactor

master/build-rust-manifest.py
```diff
@@ -0,0 +1,489 @@
...
+    while True:
+        buf = response.read(4096)
+        if not buf: break
+        f.write(buf)
+    f.close()
+
+    # Unpack the installer
+    unpack_dir = temp_dir + "/unpack"
+    os.mkdir(unpack_dir)
+    r = subprocess.call(["tar", "xzf", installer_file, "-C", unpack_dir, "--strip-components", "1"])
```
alexcrichton: I believe python has a built in `tarfile` library which can manage all this without frobbing the filesystem and shelling out, maybe not worth using though if this is already tested

master/build-rust-manifest.py
```diff
@@ -0,0 +1,489 @@
...
+    with open(local_file, 'rb') as f:
+        buf = f.read()
+        hash = hashlib.sha256(buf).hexdigest()
+
+    return {
+        "url": url,
+        "hash": hash,
+    }
+
+def write_manifest(manifest):
```
alexcrichton: If we know how to import a library like pytoml it can probably do all this for us, but if not oh well

master/build-rust-manifest.py
edunham: https://github.com/edunham/v2_rust_metadata/blob/master/v2_rust_metadata/manifest.py#L311 never seemed to introduce any major issues in my testing

master/build-rust-manifest.py
alexcrichton: (I have no idea how to import a library)

master/master.cfg
```diff
@@ -170,7 +172,7 @@ cargo_dist_platforms = [p for p in cargo_platforms if "linux" in p or "mac" in p
 
 def works_in_dev(platform):
     return 'bsd' not in platform and \
-          'bitrig' not in platform
+          'bitrig' not in platform and 'mac' not in platform
```
alexcrichton: Could this be reverted? I think we got them working right?

master/master.cfg
```diff
@@ -196,6 +198,7 @@ nogate_builders = [
 dist_nogate_platforms = [
     "cross-linux",
     "mac-ios",
+    "mac",
```
alexcrichton: Probably want to revert this before landing

master/master.cfg
```diff
@@ -1219,7 +1222,7 @@ def distsnap_buildfactory(platform, channel_label):
     elif 'cross-linux' in platform or 'ios' in platform:
         command.append('dist')
     else:
-        command.append('distcheck')
+        command.append('dist') # FIXME distcheck
```
alexcrichton: We're talking on IRC about this, but probably want to revert this? Or maybe run `check` on other platforms?

master/master.cfg
alexcrichton: Oh blocking is a killer point, disregard me!

alexcrichton:
A few blocking comments in master.cfg, but otherwise r=me for the new script and with those changes applied

master/build-rust-manifest.py
```diff
@@ -0,0 +1,489 @@
...
+    installer_url = s3_addy + "/" + dist + "/" + date + "/" + installer_name
+    print "downloading " + installer_url
+    response = urllib2.urlopen(installer_url)
+    if response.getcode() != 200:
+        raise Exception("couldn't download " + installer_url)
+
+    installer_file = temp_dir + "/" + installer_name
+    f = open(installer_file, "w")
+    while True:
+        buf = response.read(4096)
```
edunham: Tarball could get corrupted during download; is it worth checking the shasum or trying to find a local copy of the tarball (as I believe there will be for rustc though probably not cargo) first?

master/build-rust-manifest.py
edunham: `import pytoml` at the top, and making sure it's available in the environment where the buildmaster runs (`pip install pytoml` in the virtualenv. Switching to pytoml would be nice but a major refactor in order to build everything into an object for it to serialize

master/master.cfg
brson: Yes.

master/build-rust-manifest.py
brson: It could be corrupted, as could other downloads here. There aren't going to be a local rustc build available at the time this is run though without other changes to rust-buildbot. This step at least could verify the shas. At least in this case if the tarball is corrupted it probably won't unpack. If the shas are corrupted on download this script will possibly just insert garbage into the manifest.

brson:
This script has a race condition on its temporary directory when run in parallel. It's unlikely but will probably happen eventually. I think I'll add the temp directory as an argument and make them different per channel.

brson:
I've updated to address feedback. I have not changed the way tar is invoked, since this way is working, nor consolidated the platform lists, or taken any additional steps to verify downloads.

PR #72, opened 2016-03-13
linux-cross: cross toolchain for mips(el)-unknown-linux-musl

japaric:
With this in place we can use `configure --target=mips-unknown-linux-musl,mipsel-unknown-linux-musl`
and `make` to cross compile `libstd` to those two targets from within this Docker image.

Now, this change doesn't bootstrap a fully fledged cross toolchain. Instead it cross compiles the
musl libc to the target and uses the generated `musl-gcc` wrapper script as the cross compiler for
the target. This means that the "toolchain" for `mips(el)-unknown-linux-musl` has to borrow
utilities (like `ar`, `libgcc_s.so`) from other toolchain, in this case from
`mips(el)-unknown-linux-gnu`.

---

This appears to work AFAICT. I can use the cross compiled `libstd`s to build small cargo projects.
But, there's something that makes me uneasy: the produced `libstd` ends up being dynamically linked
to `mips(el)-unknown-linux-gnu`'s' `gcc_s` library which in turn is dynamically linked to `glibc`.
This means that there is a possibility that `libstd` may end with an undefined `glibc`-only symbol.
In that scenario the end user may wind up with an 'undefined reference' error when cross compiling
because this `glibc`-only symbol is not found in the musl `libc`. Now, it appears that the `gcc_s`
library contained in the image is currently free of `glibc`-isms:

```
$ mipsel-linux-gnu-nm -D -u /usr/mipsel-linux-gnu/lib/libgcc_s.so.1
         w _ITM_deregisterTMCloneTable
         w _ITM_registerTMCloneTable
         w _Jv_RegisterClasses
         w __cxa_finalize
         w __gmon_start__
         w __pthread_key_create
         U abort
         U calloc
         U dl_iterate_phdr
         U free
         U malloc
         U memcpy
         U memset
         w pthread_getspecific
         w pthread_key_create
         w pthread_mutex_lock
         w pthread_mutex_unlock
         w pthread_once
         w pthread_setspecific
         U realloc
         U strlen
```

But that may not always be the case in the future if we ever update `gcc`/`glibc`.

There's an alternative that, I believe, doesn't have this problem: Bootstrap a real cross toolchain
using crosstool-ng. In that case, the `gcc_s` library would be dynamically linked to the `musl`
libc, not to `glibc`. The cost of this alternative is that we would have to check in two more
`.config` files and the build time of the Docker image will increase.

Thoughts? @alexcrichton


alexcrichton:
Hm using `gcc_s` from MUSL definitely sounds pretty fishy and seems like something we should avoid? This is where the x86_64 target compiles its own libunwind from LLVM, would that be possible here as well?

japaric:
> Hm using gcc_s from glibc definitely sounds pretty fishy and seems like something we should avoid? 

Yeah, I agree.

> This is where the x86_64 target compiles its own libunwind from LLVM, would that be possible here as well?

We can do that but that would increase the binary size because libunwind is linked statically, and I'd like to keep binary size as low as possible for these targets because they are usually memory constrained in both RAM and flash. That being said I haven't actually measure how much difference linking to libunwind would make.

alexcrichton:
Oh right MUSL on MIPS is dynamically linked right now. What's the unwinder used at runtime for the systems you're running on? In theory we should be linking directly against that one

japaric:
> What's the unwinder used at runtime for the systems you're running on?

`gcc_s` is installed by default in OpenWRT trunk, so I guess that's what the applications use as unwinder if they use unwinding at all.

alexcrichton:
Would it be possible to put that binary in place to link against? We're not distributing so shouldn't have license problems I think.

japaric:
> Would it be possible to put that binary in place to link against? We're not distributing so shouldn't have license problems I think.

Sounds feasible. Two questions:

- How do I know for sure that `libstd` got linked against this "musl"'s `gcc_s` instead of the "glibc" one? I'm going to put the binary in musl sysroot which should have precedence over /usr/lib where glibc resides. But is there a way know *for sure* that it linked the right one?

- The gcc_s will come from OpenWRT daily snapshots (the stable version is not using musl) and there [a bunch of releases](https://downloads.openwrt.org/snapshots/trunk/) for different devices that probably have been compiled with different codegen options for performance reasons so I should try to pick the most generic of them. My question is: It shouldn't matter if gcc_s was compiled for a specific device, right? Because we are dynamic linking.

alexcrichton:
> But is there a way know for sure that it linked the right one?

Hm unfortunately I don't know of a way to do this. I thought though that the MUSL gcc wrapper would basically remove `/usr/lib` from all search paths? I'm a little surprised it's left in there...

> It shouldn't matter if gcc_s was compiled for a specific device, right?

I believe so, yeah. We're referencing the same symbols from them, so I *think* it should work out no matter what. Although along the same vein it shouldn't matter if we link to the glibc ones because you could replace them with the MUSL ones at runtime... (assuming same symbols with no weird versions in them...)

japaric:
> I thought though that the MUSL gcc wrapper would basically remove /usr/lib from all search paths?

It doesn't:

```
# Note: mips-linux-musl-gcc is the musl gcc wrapper
$ mips-linux-musl-gcc -Wl,--as-needed -lgcc_s -Wl,--verbose hello.c
(...)
attempt to open /musl-mips/lib/crt1.o succeeded
/musl-mips/lib/crt1.o <-- (*)
attempt to open /musl-mips/lib/crti.o succeeded
/musl-mips/lib/crti.o <-- (*)
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/crtbegin.o succeeded
/usr/lib/gcc-cross/mips-linux-gnu/5/crtbegin.o <-- (*)
attempt to open /musl-mips/lib/libgcc_s.so failed
attempt to open /musl-mips/lib/libgcc_s.a failed
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so succeeded
opened script file /usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so
opened script file /usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so
attempt to open libgcc_s.so.1 failed
attempt to open /musl-mips/lib/libgcc_s.so.1 failed
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1 succeeded
libgcc_s.so.1 (/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1) <--
attempt to open /musl-mips/lib/libgcc.so failed
attempt to open /musl-mips/lib/libgcc.a failed
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc.so failed
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc.a succeeded
attempt to open /tmp/ccSImUu9.o succeeded
/tmp/ccSImUu9.o
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/libgcc.a succeeded
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/libgcc_eh.a succeeded
attempt to open /musl-mips/lib/libc.so succeeded
-lc (/musl-mips/lib/libc.so)
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/libgcc.a succeeded
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/libgcc_eh.a succeeded
attempt to open /usr/lib/gcc-cross/mips-linux-gnu/5/crtend.o succeeded
/usr/lib/gcc-cross/mips-linux-gnu/5/crtend.o <-- (*)
attempt to open /musl-mips/lib/crtn.o succeeded
/musl-mips/lib/crtn.o
```

(*) I just noticed it's mixing crt1.o and crti.o from musl with crtbegin.o and crtend.o from glibc, that doesn't sound good...

And if I don't pass `-Wl,--as-needed` it actually fails because it tries to look for glibc symbols in musl:

```
$ mips-linux-musl-gcc -lgcc_s -Wl,--verbose hello.c
(...)
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `memcpy@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `calloc@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `abort@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `strlen@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `malloc@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `memset@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `realloc@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `free@GLIBC_2.0'
/usr/lib/gcc-cross/mips-linux-gnu/5/./libgcc_s.so.1: undefined reference to `dl_iterate_phdr@GLIBC_2.2.4'
```

> (assuming same symbols with no weird versions in them...)

^ I'm starting to wonder how cross compiling libstd to mips-musl actually succeded with this PR.

I still think this suggestion:

> Bootstrap a real cross toolchain using crosstool-ng.

Is the less error prone alternative.

alexcrichton:
> ^ I'm starting to wonder how cross compiling libstd to mips-musl actually succeded with this PR.

bahahaha this is great, I feel this way all the time sometimes :)

And yeah I guess if we're not controlling everything (like we do with x86_64 musl) then it may be good to have a whole toolchain to ensure we get all these references right.

alexcrichton:
Closing in favor of https://github.com/rust-lang/rust-buildbot/pull/75

PR #73, opened 2016-03-13
linux-cross: fix the `arm-linux-gnueabihf` toolchain

japaric:
The toolchain fails to build with the current settings. This PR adjusts them to make the toolchain
buildable again.

### DIFF
--- a/slaves/linux-cross/arm-linux-gnueabihf.config
+++ b/slaves/linux-cross/arm-linux-gnueabihf.config
@@ -104,7 +104,7 @@ CT_ARCH_DEFAULT_32=y
 CT_ARCH_ARCH="armv6"
 CT_ARCH_CPU=""
 CT_ARCH_TUNE=""
-CT_ARCH_FPU="vfpv3-d16"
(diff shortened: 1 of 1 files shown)


japaric:
FWIW, the `-march=armv6 -mfpu=vfp` is [what people recommend](http://raspberrypi.stackexchange.com/questions/2046/which-cpu-flags-are-suitable-for-gcc-on-raspberry-pi) for the RPi which is an ARMv6 device.

I'm not sure why thumb mode by default fails to build, but I changed it to arm mode by default.

cc @alexcrichton

japaric:
re: arm mode vs thumb mode. Looks like we are already forcing arm mode via `-marm` when [building C libraries](https://github.com/rust-lang/rust/blob/3e05371c13f851a5af57a96a5e13c99537787314/mk/cfg/arm-unknown-linux-gnueabihf.mk#L11-L12). So the change here actually makes things more consistent.

alexcrichton:
Sounds good to me!

PR #75, opened 2016-03-16
linux-cross: add mips(el)-unknown-linux-musl toolchains

japaric:
With these toolchains in place we can use the "old" build system to cross compile libstd to those
two mips targets.

---

Alternative implementation of #72 

Tested by running libc-test (cross compiled to mipsel) under a qemu emulated [OpenWRT malta](https://wiki.openwrt.org/doc/howto/qemu) device.

It should be possible to cross compile rustc to these targets using these toolchains but I haven't tested that (yet).

r? @alexcrichton

### DIFF
--- a/slaves/linux-cross/README.md
+++ b/slaves/linux-cross/README.md
@@ -110,3 +110,38 @@ For targets: `arm-unknown-linux-gnueabihf`, `armv7-unknown-linux-gnueabihf`
 (+) These options have been selected to match the gcc flags we use to compile C
     libraries like jemalloc. See the mk/cfg/arm-uknown-linux-gnueabi{,hf}.mk
     file in Rust's source code.
+
+## `mips-linux-musl.config`
+
+For targets: `mips-unknown-linux-musl`
+
+- Path and misc options > Prefix directory = /x-tools/${CT\_TARGET}
+- Target options > Target Architecture = mips
+- Target options > Endianness = **Big** endian
+- Target options > Architecture level = mips32r2 -- (+)
+- Target options > Floating point = software (no FPU) -- (+)
+- Operating System > Target OS = linux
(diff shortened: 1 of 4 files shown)


slaves/linux-cross/README.md
```diff
@@ -110,3 +110,38 @@ For targets: `arm-unknown-linux-gnueabihf`, `armv7-unknown-linux-gnueabihf`
...
+- Target options > Target Architecture = mips
+- Target options > Endianness = **Big** endian
+- Target options > Architecture level = mips32r2 -- (+)
+- Target options > Floating point = software (no FPU) -- (+)
+- Operating System > Target OS = linux
+- Operating System > Linux kernel version = 4.3 -- OpenWRT trunk uses 4.4 though
+- C-library > C library = musl
+- C-library > musl version = 1.0.5 -- OpenWRT trunk uses 1.1.14 though
+- C compiler > gcc version = 5.2
+- C compiler > C++ = ENABLE -- compiler-rt demands a C++ compiler even when we don't use it
```
alexcrichton: grrr compiler-rt

I guess there's no way we can hack around this in the build of compiler-rt? We shouldn't in theory need to produce any MIPS Rust compilers right now, right?

alexcrichton:
Nice!

slaves/linux-cross/README.md
japaric: Wait, I think this only necessary when using rustbuild and we are going to use the old build system to cross compile libstd to these mips targets. IOW I need to double check this.

japaric: (I guess the downside of having two build systems is that I'm mixing up their requirements...)

alexcrichton: Hehe oh well, can't really hurt to have a C++ compiler at the ready I guess

PR #77, opened 2016-03-16
linux-cross: add armv7-linux-gnu toolchain

japaric:
When using rustbuild to build libstd/rustc for the armv7-unknown-linux-gnueabihf target we use this
toolchain instead of the arm-linux-gnueabihf one.

This commit also installs cmake and gives a home directory to the rustbuild user. Both are required
to be able to use rustbuild to cross compile rustc within this image.

r? @alexcrichton

### DIFF
--- a/slaves/linux-cross/Dockerfile
+++ b/slaves/linux-cross/Dockerfile
@@ -80,6 +80,19 @@ RUN /bin/bash build_toolchain.sh arm-linux-gnueabihf
 RUN /bin/bash build_toolchain.sh mips-linux-musl
 RUN /bin/bash build_toolchain.sh mipsel-linux-musl
 
+# Also build a toolchain tuned for the armv7 architecture which is going to be
+# used with the armv7-unknown-linux-gnueabihf target.
+#
+# Why are we not using the arm-linux-gnueabihf toolchain with the armv7 target?
+# We actually tried that setup but we hit `ar` errors caused by the different
+# codegen options used by crosstool-ng and the rust build system. crosstool-ng
+# uses `-march=armv6` to build the toolchain and related C(++) libraries, like
+# libstdc++ which gets statically linked to LLVM; on the other hand the rust
+# build system builds its C(++) libraries, like LLVM, with `-march=armv7-a`.
+#
+# By using this armv7 compiler we can ensure the same codegen options are used
+# everywhere and avoid these codegen mismatch issues. Also compiling libstdc++
+# for armv7 instead of for armv6 should make rustc (slightly) faster.
 RUN /bin/bash build_toolchain.sh armv7-linux-gnueabihf
 
 USER root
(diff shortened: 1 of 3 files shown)


slaves/linux-cross/Dockerfile
```diff
@@ -79,6 +80,8 @@ RUN /bin/bash build_toolchain.sh arm-linux-gnueabihf
 RUN /bin/bash build_toolchain.sh mips-linux-musl
 RUN /bin/bash build_toolchain.sh mipsel-linux-musl
 
+RUN /bin/bash build_toolchain.sh armv7-linux-gnueabihf
```
alexcrichton: Could you add a comment here explaining why we build a whole new toolchain instead of just using the `arm-linux-gnueabihf` one above with different codegen flags? (the reason being statically linking to libstdc++ for the compiler we'll ship)

slaves/linux-cross/Dockerfile
```diff
@@ -101,8 +104,18 @@ RUN                                               \
...
+  for f in `ls /x-tools/armv7-unknown-linux-gnueabihf/bin/armv7-unknown-linux-gnueabihf-*`; do     \
+    g=`basename $f`;  \
+    ln -vs $f /usr/bin/`echo $g | sed -e 's/-unknown//'`;    \
   done
 
+# Instruct rustbuild to use the armv7-linux-gnueabihf toolchain instead of the default
+# arm-linux-gnueabihf one
+ENV AR_armv7_unknown_linux_gnueabihf=armv7-linux-gnueabihf-ar \
+  CC_armv7_unknown_linux_gnueabihf=armv7-linux-gnueabihf-gcc \
+  CXX_armv7_unknown_linux_gnueabihf=armv7-linux-gnueabihf-g++
```
alexcrichton: Hm this is an interesting question, I'm not sure if it's best to put this configuration here or in the buildbot config. I guess for now it seems fine to put here? Perhaps this could go ahead and add environment variables for basically all the cross triples?

japaric:
Added the requested comment

slaves/linux-cross/Dockerfile
japaric: > Perhaps this could go ahead and add environment variables for basically all the cross triples?

We don't really *need* to set all the env variable for all the cross triple because the default toolchains chosen by the gcc crate(?) match the toolchains installed in this docker image. Unless you want to be explicit about it. In any case, where can I get the full list of cross triples we are currently building binaries for?

alexcrichton: Yeah the choices of the gcc crate are in theory made to reflect whatever the "standard toolchain" is named for any cross target, but unfortunately most of these targets seem like they don't necessarily have standard toolchains. Right now the "standard" is whatever ubuntu happens to have.

And eh you're right, we'll pick up the others by default anyway. I think the MIPS-MUSL ones won't work by default (but gcc doesn't have a case for that yet), so seems fine to add over time. I'll go ahead and merge!

PR #79, opened 2016-03-17
linux: build musl/libunwind for i686

japaric:
Now we can use the "linux" docker image to build libstd for i686-unknown-linux-musl with the
following command:

configure --disable-rustbuild --musl-root=/musl-i686 --target=i686-unknown-linux-musl && make

r? @alexcrichton

### DIFF
--- a/slaves/linux/build-musl.sh
+++ b/slaves/linux/build-musl.sh
@@ -31,7 +31,8 @@ make -j10
 cp lib/libunwind.a /musl-x86_64/lib
 
 # (Note: the next cmake call doesn't fully override the previous cached one, so remove the cached
-# configuration manually)
+# configuration manually. IOW, if don't do this or call make clean we'll end up building libunwind
+# for x86_64 again)
 rm -rf *
 # for i686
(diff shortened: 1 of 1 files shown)


slaves/linux/build-musl.sh
```diff
@@ -7,9 +7,15 @@ export CFLAGS=-fPIC
 # Support building MUSL
 curl http://www.musl-libc.org/releases/musl-1.1.11.tar.gz | tar xzf -
 cd musl-1.1.11
+# for x86_64
 ./configure --prefix=/musl --disable-shared
```
alexcrichton: Could you change this to install to `/musl-x86_64`?

alexcrichton:
Nice! I think we may want to change this at some point so you can configure the i686 and x86_64 musl target at the same time, but right now they both try to use `musl-gcc` which fails. Something we can fix later though!

japaric:
> Could you change this to install to /musl-x86_64?

Done. Don't forget to update the `--musl-root` flag on the rustbuild side!

alexcrichton:
pffffft thinking we don't test changes before we merge them

I will indeed attempt to remember :)

PR #81, opened 2016-03-19
Lotsa compilers

alexcrichton:
These commits do a few things:

* Three new auto builders are created:
  * `linux-64-x-netbsd` - tests cross compiling from Linux to NetBSD
  * `linux-64-x-freebsd` - tests cross compiling from Linux to FreeBSD
  * `linux-64-x-armsf` - tests cross compiling from x86_64 linux to arm-*-*-gnueabi (soft float) linux
  * `linux-64-x-armhf` - tests cross compiling form x86_64 linux to arm-*-*-gnueabihf (hard float) linux
* One new dist builder is created:
  * `cross-host-linux` - produces host compilers

The goal of this PR is to prep both our gated commits as well as our nightlies
for producing NetBSD, FreeBSD, ARM, ARMv7, and ARM (hard float) compilers. Some
tinings of this were tested in dev with respect to the time it takes for an auto
builder to compile the number of *extra* hosts. In the table below one column
is bootstrapping plus compiling LLVM (a cold cache) and another is just the
bootstrap where we didn't have to recompile LLVM (a warm cache)

| Extra hosts | boostrap + LLVM |   boostrap   |
|-------------|-----------------|--------------|
|     2       |   3hrs 4min     |  2hrs 7min   |
|     3       |   4hrs 24min    |  2hrs 55min  |
|     5       |   6hrs 56min    |     N/A      |

Note that there was some contention on the machines so that's why the drops
aren't exactly as one might expect. In any case it looks like we shouldn't
bootstrap more than *one* extra host on an auto builder to keep build times
reasonable. As a result, each of the `-x-` builders added only crosses to one
extra host, and a builder wasn't added for ARMv7 as it should be the same as ARM
hard float.

Hopefully once this lands and we update the builders, we'll just magically get a
whole lot more compilers overnight!

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -656,6 +656,8 @@ class MakeCommand(object):
     implements(IRenderable)
     def getRenderingFor(self, props):
         if "buildername" in props:
+            if "linux" in props["buildername"]:
+                return "make"
             if "bsd" in props["buildername"] or "bitrig" in props["buildername"]:
                 return "gmake"
             if "win" in props["buildername"]:
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -1727,6 +1727,7 @@ for p in dist_platforms:
 
     for channel in ['nightly', 'beta', 'stable']:
         my_targets = targets[:]
+        my_hosts = hosts
         rustbuild = None
         if "cross-linux" in p:
             if channel == 'beta':
@@ -1735,7 +1736,7 @@ for p in dist_platforms:
                 rustbuild = True
                 my_targets += nightly_lincross_targets
                 my_targets += beta_lincross_targets
-                hosts = [
+                my_hosts = [
                   "arm-unknown-linux-gnueabi",
                   "arm-unknown-linux-gnueabihf",
                   "armv7-unknown-linux-gnueabihf",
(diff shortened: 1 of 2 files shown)


alexcrichton:
Hm so one question that I'm remembering now, what happens if we produce the same standard library tarball on two hosts? The new `cross-host-linux` builder overlaps with the `cross-linux` builder a bit as they'll both produce `rust-std-nightly-arm-unknown-linux-gnueabihf.tar.gz` (for example). I can fix this, but it may be a pain.

@brson do you know if this will cause problems? If one just overwrites the other we can probably just leave it maybe?

alexcrichton:
cc @japaric

brson:
I want to say that it will fail if there are artifacts from multiple builders with the same name, but I don't see in the code why that would be the case. Let's just land it and see if nightlies break.

master/master.cfg
```diff
@@ -1710,12 +1734,23 @@ for p in dist_platforms:
...
                 my_targets += nightly_lincross_targets
                 my_targets += beta_lincross_targets
+        if "cross-host-linux" in p:
+            rustbuild = True
+            my_hosts = [
+              "arm-unknown-linux-gnueabi",
+              "arm-unknown-linux-gnueabihf",
+              "armv7-unknown-linux-gnueabihf",
+              "x86_64-unknown-netbsd",
+              "x86_64-unknown-freebsd",
```
brson: Can you extract this out like the other `*_lincross_targets` lists?

brson:
r=me

PR #89, opened 2016-03-30
slaves/linux: install newer cmake

japaric:
required to build cargo on this image

follow-up of rust-lang/cargo#2532

r? @alexcrichton 

I didn't add /musl-x86_64/bin to PATH in the Dockerfile. I think it would be better to do that in Cargo's configure script because this image is going to be used to cross compile std to x86_64-musl and i686-musl where both tasks also use `musl-gcc` so we may accidentally end up using `/musl-x86_64/bin/musl-gcc` when building std for i686-musl. Also if we ever build cargo for i686-musl, we'll probably end up using Cargo's configure script to pick the right `musl-gcc` anyways.

### DIFF
--- a/slaves/linux/Dockerfile
+++ b/slaves/linux/Dockerfile
@@ -3,10 +3,16 @@ FROM ubuntu:14.04
 RUN dpkg --add-architecture i386
 RUN apt-get update
 RUN apt-get install -y \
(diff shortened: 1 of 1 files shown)


alexcrichton:
Ok, I've pushed this under the tag 2016-03-30

PR #94, opened 2016-04-15
linux-cross: install toochain for 32-bit freebsd

japaric:
I can `configure --enable-rustbuild --host=i686-unknown-freebsd && make` with this change.

r? @alexcrichton


alexcrichton:
Nifty!

PR #100, opened 2016-04-25
update Android builds to use NDK r11c

froydnj:
One of the changes in Android NDK r11 and above is:

  Removed from all versions of NDK libc, m, and dl all symbols which the
  platform versions of those libs do not support.

This cryptic statement means that a number of symbols that the header
files no longer use were removed from libc and friends.  What this means
in practice is that the Rust libraries for arm-linux-androideabi contain
references to symbols that no longer exist in current versions of the
NDK.  Which means that you have to use the exact version of the NDK (or
previous versions, presumably) that Rust was built with to use the
prebuilt binaries, but you can't use any later versions.

This situation seems undesirable.  To alleviate this, this patch moves
to using version r11c of the NDK, which ought to be just as
backwards-compatible, as well as (more) forwards-compatible.  Moving to
r11c requires that the arm-linux-androideabi target now use GCC 4.9, but
that should not be a problem, as other Android architectures Rust builds
for were already using GCC 4.9.

### DIFF
--- a/slaves/android/install-ndk.sh
+++ b/slaves/android/install-ndk.sh
@@ -5,38 +5,37 @@ set -ex
 # Prep the Android NDK
 #
 # See https://github.com/servo/servo/wiki/Building-for-Android
-curl -O http://dl.google.com/android/ndk/android-ndk-r10e-linux-x86_64.bin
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice!

PR #102, opened 2016-05-21
Include source packages in manifest

Diggsey:
This will allow rustup to install the rust source as an optional extension for a given toolchain.

### DIFF
--- a/master/build-rust-manifest.py
+++ b/master/build-rust-manifest.py
@@ -295,12 +295,16 @@ def build_manifest(rustc_date, rustc_version, rustc_short_version,
     mingw_pkg = build_package_def_from_archive("rust-mingw", "dist", rustc_date,
                                                rustc_version, rustc_short_version,
                                                mingw_list)
+    src_pkg = build_package_def_from_archive("rust-src", "dist", rustc_date,
+                                             rustc_version, rustc_short_version,
+                                             ["*"])
 
     packages["rustc"] = rustc_pkg
     packages["rust-std"] = std_pkg
     packages["rust-docs"] = doc_pkg
     packages["cargo"] = cargo_pkg
     packages["rust-mingw"] = mingw_pkg
+    packages["rust-src"] = src_pkg
 
     # Build the rust package. It is the only one with subcomponents
     rust_target_pkgs = {}
@@ -331,6 +335,12 @@ def build_manifest(rustc_date, rustc_version, rustc_short_version,
                 "pkg": "rust-std",
                 "target": target,
             }]
+
+        # The src package is also an extension
+        extensions += [{
+            "pkg": "rust-src",
+            "target": "*",
+        }]
 
         # The binaries of the 'rust' package are on the local disk.
         # url_and_hash_of_rust_package will try to locate them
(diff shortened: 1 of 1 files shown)


Diggsey:
The source packages are a little different from other packages, in that they are not in the usual "installer" format (they don't have a components file describing how to install the package).

From rustup's POV, I plan on treating packages which don't contain a components file as "pure data" packages, which will just get copied to a suitably named subdirectory within the toolchain subdirectory.

alexcrichton:
Thanks for the PR! I suspect that @brson has many thoughts on this, so I'm somewhat hesitant to deploy a new format in the interim before he weighs in. I personally feel somewhat odd about using "*" as the "target" here, but so long as it works it's just an implementation detail so I'm not too worried.

rillian:
Something like this is important I think to support the IDE integration tools which need copies of libcore and libstd for indexing. Having the installer put the source files in $prefix/src/ would make the sources easier to find.

alexcrichton:
Ah yeah we definitely want to include them! This is just a question of how we modify our metadata to include info about where to download them from.

brson:
Thanks for moving on this.

It looks to me like this is including the existing source tarball in the manifest, which is not in the rust-installer format. If we were to go that route then clients interpreting the manifest would have to know that tarball is special, and I prefer not to do that. So before we do this we need to modify the rust build system to produce a new source artifact that is built by rust-installer, so that rustup can treat it like any other installer.

The current source tarball is called "rustc-$version-src.tar.gz", and we'll want the build system to *additionally* produce a rust-installer tarball called "rustc-src-$version.tar.gz".

As to the installation location, it should probably be `$prefix/lib/rustlib/src`, and that will need to be configured when the rust-installer package is built in the rust build system.

The question of what to do with the target name is tricky. Fortunately rust-installer itself doesn't know anything about targets - it's just part of the package name. rustup though does care about targets and expects every component to have a name/target pair, so it will need to be extended to deal with components that don't have a target. I *think* using an asterisks here will be fine but I'm not sure of all the details it will affect.

Diggsey:
I'd prefer to use `rust-src` or similar rather than `rustc-src` since this contains both the source for the standard library and rustc.

This shouldn't cause much of a problem for rustup: the modifications (if any are even needed) should be fairly self contained as it only uses the package targets to immediately lookup that package within the manifest.

I chose to use asterisks because it required the minimal number of changes, so made sense to do that until I could get some feedback, but modifying the manifest format also works.

brson:
Oh, yes I prefer `rust-src` too. Typo.

liuchong:
will it also set a "RUST_SRC_PATH" env var? racer and other apps are using this.

retep998:
@liuchong That would probably be handled by a combination of rustup.rs and whatever editor tooling you're using. This PR is just allowing for source packages to be created on the buildbots.

master/build-rust-manifest.py
```diff
@@ -402,8 +412,14 @@ def live_package_url(name, dist_dir, date, version, target):
     if name == "cargo":
         maybe_channel = "nightly"
 
-    url1 = s3_addy + "/" + dist_dir + "/" + date + "/" + name + "-" + version + "-" + target + ".tar.gz"
-    url2 = s3_addy + "/" + dist_dir + "/" + date + "/" + name + "-" + maybe_channel + "-" + target + ".tar.gz"
+    if name == "rust-src":
+        # The build system treats source packages as a separate target for `rustc`
+        # but for rustup we'd like to treat them as a completely separate package.
+        url1 = s3_addy + "/" + dist_dir + "/" + date + "/rustc-" + version + "-src.tar.gz"
+        url2 = s3_addy + "/" + dist_dir + "/" + date + "/rustc-" + maybe_channel + "-src.tar.gz"
```
brson: @Diggsey It looks to me like the only thing that needs to change is the naming scheme here. It's `rust-src-$version-or-channel.tar.gz` now.

Diggsey:
Fixed up the name :grinning:

brson:
I'm going to test this in the dev environment to make sure existing rustup deployments don't break on the new metadata. I suspect there are some tweaks needed still. Starting a build in dev now.

brson:
The only thing that needs to change is that `[pkg.rust-src.target.*]` needs to output as `[pkg.rust-src.target."*"]`.

PR #104, opened 2016-05-25
Openssl 1.0.2h

rillian:
A couple of updates to the buildbot base. The critical openssl update doesn't affect us (privacy-impact only) but there are a couple of overflow fixes which could. The git and curl bumps are just updating to the latest.

I expect we'll want to wait until after tomorrow's merge day to deploy these, but I did the bump on my gecko-side builder.

r? @edunham


alexcrichton:
Thanks!

PR #105, opened 2016-05-27
Add DragonFly toolchain

mneumann:
This sets up the toolchain (binutils, gcc, libraries) needed for cross-compiling to DragonFly.
Note that `bsdtar` is required to unpack the DragonFly ISO image, as there are no .tar.gz releases of DragonFly.

### DIFF
--- a/slaves/linux-cross/build_dragonfly_toolchain.sh
+++ b/slaves/linux-cross/build_dragonfly_toolchain.sh
@@ -58,6 +58,16 @@ cd gcc
 curl https://ftp.gnu.org/gnu/gcc/gcc-$GCC/gcc-$GCC.tar.bz2 | tar xjf -
 cd gcc-$GCC
 
+# The following three patches are taken from DragonFly's dports collection:
+# https://github.com/DragonFlyBSD/DPorts/tree/master/lang/gcc5
+# The dports specification for gcc5 contains a few more patches, but they are
+# not relevant in this situation, as they are for a language we don't need
+# (e.g. java), or a platform which is not supported by DragonFly (e.g. i386,
+# powerpc64, ia64, arm).
+#
+# These patches probably only need to be updated in case the gcc version is
+# updated.
+
 patch -p0 <<'EOF'
 --- libatomic/configure.tgt.orig	2015-07-09 16:08:55 UTC
 +++ libatomic/configure.tgt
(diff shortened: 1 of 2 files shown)


slaves/linux-cross/build_dragonfly_toolchain.sh
```diff
@@ -0,0 +1,125 @@
...
+rm -rf dragonfly
+
+# Finally, download and build gcc to target FreeBSD
+mkdir gcc
+cd gcc
+curl https://ftp.gnu.org/gnu/gcc/gcc-$GCC/gcc-$GCC.tar.bz2 | tar xjf -
+cd gcc-$GCC
+
+patch -p0 <<'EOF'
+--- libatomic/configure.tgt.orig	2015-07-09 16:08:55 UTC
```
alexcrichton: Do you have links for where these patches come from? Just so if we need to update this we know where to look

mneumann: they are from DragonFly's dports collection (dports =~ FreeBSD ports): https://github.com/DragonFlyBSD/DeltaPorts/tree/master/ports/lang/gcc5/dragonfly or https://github.com/DragonFlyBSD/DPorts/tree/master/lang/gcc5. DeltaPorts is our layer above FreeBSD's ports collection. DPorts is then created by applying the patches from DeltaPorts against FreeBSD's ports.

mneumann: I think you only need to update the patches in case you change the gcc version. Newer versions of gcc might already ship with those patches.

alexcrichton: Thanks for the info! Could you add that as a comment here? We may need to update gcc one day and it'd just be good to know where the patches came from so we could perhaps pick up others if they're necessary.

mneumann:
@alexcrichton Have added a comment.

alexcrichton:
Thanks!

To confirm as well, have you built a toolchain inside of the container and confirmed that it works locally on DragonFly as well?

mneumann:
@alexcrichton I can confirm that I can cross-compile a DragonFly executable from within the container, which later can be executed on a recent DragonFly system. I only tested a simple hello world application written in C. For this, ```file``` returns:
```
/tmp/t: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /usr/libexec/ld-elf.so.2, for DragonFly 4.0.400, not stripped
```

What I haven't done yet is to build a rust toolchain from within the container. This sounds a little bit more complicated.

PR #106, opened 2016-06-20
Reduce the number of auto builds

alexcrichton:
* mac-64-nopt-t - do nopt-t on Linux
* linux-32cross-opt - rarely turns up regressions and we have nightlies
* linux-64-cross-netbsd - rarely turns up regressions and we have nightlies
* linux-64-cross-armsf - rarely turns up regressions and we have nightlies
* linux-64-cross-armhf - rarely turns up regressions and we have nightlies
* win-gnu-32-nopt-t - do nopt-t on Linux
* win-gnu-64-nopt-t - do nopt-t on Linux
* win-msvc-32-cross-opt - rarely turns up regressions and we have nightlies
* bitrig/freebsd/dragonfly/openbsd - bots haven't been online in awhile

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -112,7 +112,7 @@ auto_platforms = [
     #"mac-32-nopt-t",
     "mac-64-opt",
     #"mac-64-nopt-c",
-    "mac-64-nopt-t",
+    #"mac-64-nopt-t",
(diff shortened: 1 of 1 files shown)


brson:
:(

PR #108, opened 2016-07-19
switch mipsel-musl to soft float

TimNN:
r? @alexcrichton 

cc rust-lang/rust#34922


slaves/linux-cross/README.md
```diff
@@ -165,6 +165,7 @@ For targets: `mipsel-unknown-linux-musl`
 - Target options > Target Architecture = mips
 - Target options > Endianness = **Little** endian
 - Target options > Architecture level = mips32 -- (+)
+- Target options > Floating point = software (no FPU) -- (+)
```
TimNN: Not sure whether or not that `(+)` is correct here, this line was basically copied from the `mips-unknown-linux-musl` target.

alexcrichton:
Thanks!

cc @japaric

PR #109, opened 2016-07-27
prepare for binary releases of std for arm-musl

japaric:
this commit adds toolchains for the following targets:

- arm-unknown-linux-musleabi
- arm-unknown-linux-musleabihf
- armv7-unknown-linux-musleabihf

because cross compiling std for these targets requires linking to a
cross compiled libunwind, this commit also cross compiles libunwind

---

r? @alexcrichton 
cc @timonvo

### DIFF
--- a/slaves/linux-cross/Dockerfile
+++ b/slaves/linux-cross/Dockerfile
@@ -51,10 +51,13 @@ RUN /bin/bash build_toolchain_root.sh && \
 COPY linux-cross/build_toolchain.sh \
     linux-cross/aarch64-linux-gnu.config \
     linux-cross/arm-linux-gnueabi.config \
+    linux-cross/arm-linux-musleabi.config \
     linux-cross/arm-linux-gnueabihf.config \
+    linux-cross/arm-linux-musleabihf.config \
     linux-cross/mips-linux-musl.config \
     linux-cross/mipsel-linux-musl.config \
     linux-cross/armv7-linux-gnueabihf.config \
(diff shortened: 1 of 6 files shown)


alexcrichton:
I wonder if we could disable all the C++ compilers here? Or do you think we should ship musl versions of the compiler here?

compiler-rt's old dependency on a C++ compiler is now gone, so we may be able to trim that part of the build.

alexcrichton:
And as always, looks great to me, thanks!

japaric:
> I wonder if we could disable all the C++ compilers here?

I was wondering the same. Ultimately, I left them in because if, in the future, we want to build rustc for arm-musl then we won't need to update the image :-).

> Or do you think we should ship musl versions of the compiler here?

I think that may not be feasible right now. I haven't tested cross compiling rustc for these targets. But, given the precedence of rustc for "static" x86-musl not working (segfault in llvm), it's probable that this won't work out of the box either. If that's the case then "dynamic" arm-musl targets may be needed to be able to build a rustc that works.

We can drop the C++ compilers. I don't have strong opinion about keeping them or dropping them, to be honest.

alexcrichton:
Eh seems reasonable to me. My only hesitation is that it's quickly turning into a 24 hour build time to build this image, but I guess I don't build it that often so it's not *that* relevant

alexcrichton:
Ok, image is built and is now pushing (may take awhile)

alexcrichton:
Now deployed!

PR #110, opened 2016-08-01
turn off newer RELX relocation optimizations

m4b:
otherwise breaks the link step on systems with an older linker (i.e., travis CI, some ubuntu's, etc.). ref rust-lang/rust#34978

We may only want to add this flag on nightly? Not sure how to do that...

*NB* requires a newer assembler (don't know exact minimal working version), but
```
as -version
GNU assembler (GNU Binutils) 2.26.0.20160501
```
works.

### DIFF
--- a/slaves/linux/build-musl.sh
+++ b/slaves/linux/build-musl.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-export CFLAGS=-fPIC
+export CFLAGS=-fPIC -Wa,-mrelax-relocations=no
 MUSL=musl-1.1.14
 # Support building MUSL
 curl http://www.musl-libc.org/releases/$MUSL.tar.gz | tar xzf -


alexcrichton:
Thanks! We shouldn't have to worry about the assembler version here as we're in control of it, but I think we'll also need to pass this to the build system, right? I remember the error being with jemalloc as well I think...

alexcrichton:
Ok, I've pushed this as alexcrichton/rust-slave-linux:2016-08-01, it should go out the next time we restart the buildbots. Do you want to send a patch to set this for the musl target as well?

m4b:
@alexcrichton I thought this was the musl target?

and yea, will need to pass to the build system, as the relocs show up in `libstd` and `liballoc_jemalloc`, as well as `libunwind` (but I hope it's cmake will obey the `CFLAGS` var?).

Unfortunately I'm not familiar with that part of building rust, and don't know if we have an equivalent `CFLAGS` somewhere in the pipeline?

alexcrichton:
Yeah it should be easy enough to modify the `mk/cfg/*.mk` files and also just add custom flags in `src/bootstrap` for the rustbuild build system. Also yeah this should only affect musl, and I think in the build system we only need to really worry about jemalloc, other pieces are fine to ignore I think.

alexcrichton:
Ok, new image is now deployed

PR #112, opened 2016-08-04
add CXXFLAGS for cmake

m4b:
Because it doesn't seem to respect CFLAGS, and we still have RELX relocations in `libunwind.a` :(

ref rust-lang/rust/issues/34978


alexcrichton:
Ok pushed as alexcrichton/rust-slave-linux:2016-08-04, will deploy on the next buildbot restart

alexcrichton:
Ok, deployed!

PR #114, opened 2016-08-07
Add DragonFly builder

mneumann:


master/master.cfg
```diff
@@ -263,6 +263,7 @@ stable_cross_targets = [
   {'t': 'x86_64-apple-ios', 'b': ios},
   {'t': 'x86_64-rumprun-netbsd', 'b': lincross},
   {'t': 'x86_64-unknown-freebsd', 'b': xhost('freebsd'), 'host': True, 'pkg': True},
+  {'t': 'x86_64-unknown-dragonfly', 'b': xhost('dragonfly'), 'host': True, 'pkg': True},
```
alexcrichton: I think that `b` here indicates that we have an auto builder with the name "dragonfly" in it to gate all changes to master on, but for now this PR doesn't add a corresponding auto builder. That's fine and we probably want to stick to that route for now, we'll just assume that dragonfly doesn't break that often (b/c breakage will break nightlies).

Also can you add this target to the nightlies array? We can then let it ride the trains back down to stable.

edunham:
These configs are no longer used. New builders can be added via PR to the .travis.yml at https://github.com/rust-lang/rust/blob/master/.travis.yml.

PR #119, opened 2016-08-24
Remove no-mir bots.

eddyb:
~~Needed to merge~~ Not needed anymore after rust-lang/rust#35764.


alexcrichton:
We'll pick this up on the next buildbot restart, but that may be awhile

PR #123, opened 2016-09-19
cross/arm: bump glibc to 2.16

japaric:
binaries built using the glibc-2.14 based toolchain segfaulted vhen
executed on ARM hardware (see issue linked below) due to a glibc bug.
Bumping the glibc version fixes the issue.

fixes rust-lang/rust#35843

r? @alexcrichton 

I've tested that the glibc-2.16 based toolchain produces good binaries for the armv7 target. I also tried to build a glibc-2.15 based toolchain but crosstool-ng failed to build/compile it.


alexcrichton:
Thanks!

PR #125, opened 2016-09-22
openssl 1.0.2i

rillian:
New release addressing a number of security issues. At first glance, risk to our application is low (denial of service). Still important to upgrade of course. Issue list at https://www.openssl.org/news/secadv/20160922.txt

Also, the old source package has been taken down, so updating is necessary for our docker images to bootstrap.


alexcrichton:
Thanks!

PR #129, opened 2016-10-03
Add emscripten targets to new 'rustbuildcross' slaves

brson:
I'm still waiting for the nightly to finish in dev, but I think it will be successful.

I ended up doing this by adding new "rustbuildcross" slaves, which of course will take up more resources...


brson:
This doesn't actually work correctly....

brson:
I'm going to force-push a revert.

PR #130, opened 2016-10-04
Add emscripten targets to new 'rustbuild-cross' slaves

brson:
OK this one definitely works. Needs https://github.com/rust-lang/rust/pull/36944 before it will actually work.

r? @alexcrichton


brson:
Oh hm let me rebase on the previous.

brson:
Rebased, and now needs to be tested again...

alexcrichton:
I really wish we version controlled `slave-lists.txt`  ...

PR #137, opened 2016-11-11
Curl 7.51.0

rillian:
This is an important security update for curl. The cmake bump is just along for the ride.


alexcrichton:
Ah yes, thanks!

Issue #140, opened 2016-12-08
Find command errors in `SetPropertyFromCommand` for property "freshconfig" on builder `auto-win-msvc-64-opt-rustbuild`


DirkyJerky:
On an unrelated build: https://buildbot.rust-lang.org/builders/auto-win-msvc-64-opt-rustbuild/builds/3217
The """freshconfig""" key is set to the value """0 /usr/bin/find: paths must precede expression: ^ Usage: /usr/bin/find [-H] [-L] [-P] [-Olevel] [-D help|tree|search|stat|rates|opt|exec] [path...] [expression]"""

Relevant source: https://github.com/rust-lang/rust-buildbot/blob/master/master/master.cfg#LC976

The find command is valid on my macOS machine here.

It caused the `should_wipe` function to return the wrong value, making the `WORK_DIR` directory to not be removed before `git clean -d -f -f` in the next step. No clue if it contributed to the build failure.

alexcrichton:
auto bots are on there way out, so closing.

PR #142, opened 2016-12-14
Change how `freshconfig` property is set

DirkyJerky:
... to make it more cross-platform friendly.

https://github.com/rust-lang/rust-buildbot/issues/140


edunham:
Thanks for the fix! However, we've moved off of Buildbot and are now using Travis. If you run into similar issues over there, please take a look at https://github.com/rust-lang/rust/blob/master/.travis.yml.

PR #143, opened 2016-12-14
Add a Fuchsia build slave

raphlinus:
Adds a Docker script to build an image containing toolchain and sysroot for Fuchsia operating system.

### DIFF
--- a/slaves/fuchsia/Dockerfile
+++ b/slaves/fuchsia/Dockerfile
@@ -1,0 +1,76 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -y --force-yes --no-install-recommends \
+	build-essential \
+	ca-certificates \
+	curl \
+	git \
+	golang \
+	unzip
+
+RUN curl -s https://raw.githubusercontent.com/fuchsia-mirror/jiri/master/scripts/bootstrap_jiri | bash -s fuchsia
+WORKDIR /fuchsia
+ENV PATH="/fuchsia/.jiri_root/bin:${PATH}"
+
+RUN jiri import sysroot https://fuchsia.googlesource.com/manifest
+
+# Patch the manifest to add scripts and llvm (for Fuchsia.cmake), and
+# also to bring in tools from magenta-rs. Hopefully the need for this
+# will go away soon.
+COPY fuchsia/manifest-patch /fuchsia/
+RUN patch .jiri_manifest < manifest-patch
+
+# Note: we have to fix the file ownership here or else it will cause
+# permissions problems when we change USER to rustbuild
+RUN jiri update && chmod -R u+rwX,go+rX /fuchsia/buildtools/toolchain
+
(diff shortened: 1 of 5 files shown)


raphlinus:
r? @alexcrichton 

I think I've got it working right (the build-head-rust.sh script does build a working compiler) but I'm sure there's lots to improve - this is the first time I've ever tried to make a Docker script.

Also @petrhosek might want to take a look.

alexcrichton:
Awesome, thanks! 

Could this logic go into the `linux-cross` image like the other toolchains as well? From an infra side at least it'll be easier to just throw everything into one for now while we're still working out how to split them apart. Also docker tends to have smaller images with fewer `RUN` steps as it saves the state between each step, so could the logic here be extracted to a `build_fuchsia_toolchain.sh` like the existing `build_freebsd_toolchain.sh` and such?

I don't mind doing these changes as well if things are getting to confusing too :)

raphlinus:
Yes. @petrhosek took a look and has some ideas how to make it lighter weight, downloading less stuff (so direct download of blobs and git repos, not using jiri). I think we can shell'ify and move it into linux-cross as part of that work.

alexcrichton:
Ok cool, sounds good to me! Don't worry too much about binary size though, it's a pretty minor concern. The `linux-cross` image literally takes hours to build, and is gigabytes in size, so we're not optimizing much here.

alexcrichton:
Oh also feel free to just ping me whenever this is updated and I'll take another look.

antiagainst:
@raphlinus: Instead of basing everything on Ubuntu, [Alpine](https://hub.docker.com/_/alpine/) is more lightweight. But I see other slaves are also using Ubuntu, which is kinda the de facto Linux distro and more widely used. So please just ignore me if the issue was already raised and explained.

petrhosek:
@alexcrichton All the important logic is going to be encapsulated in a shell script so it could be definitely used from linux-cross. By the way, I've noticed that for all other platforms, you're building the host toolchain from source (i.e. GCC), is that preferred to using prebuilt binary? I'm perfectly happy to build Clang/LLVM from source, but it can take some 20-30 minutes to build on my machine. On the other hand, if the container already takes hours to build, it probably doesn't make much of a difference.

alexcrichton:
@antiagainst ah yeah the ship sailed on "small images" long ago when we decided to install ~10 gcc toolchains on one machine :)

@petrhosek oh we just compile gcc from source to compile an older glibc. We like the binaries we distribute to be "as compatible as possible" which necessitates compiling against older glibc binaries.

For Fuchsia, though, there should be no need for that, so either a prebuilt binary or building from source works just fine! I'd almost prefer prebuilt b/c it means building the image is that much faster :)

edunham:
@raphlinus Thanks for all the work on this! We've moved our build system from BuildBot to Travis, so if this is a change that you'd still like to see land, please PR it over at https://github.com/rust-lang/rust/blob/master/.travis.yml.

PR #147, opened 2017-01-03
Python 2.7.13.

rillian:
Also move to the tar.xz package, which is a 27% smaller.

### DIFF
--- a/slaves/dist/build_python.sh
+++ b/slaves/dist/build_python.sh
@@ -2,19 +2,19 @@
 
 set -ex
 
-VERSION=2.7.12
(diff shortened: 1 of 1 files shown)


rillian:
Also note that if you haven't rebuilt the builtbot containers recently, please check that you've got the fix for https://rhn.redhat.com/errata/RHSA-2016-2962.html. I had to do `docker build --no-cache=true` to re-run the `dnf update` step; the `docker.io/centos:5` image doesn't appear to have it.

alexcrichton:
Thanks!

PR #150, opened 2017-01-25
Add rust-analysis manifest

nrc:
r? @brson


alexcrichton:
I think we'll also want to add this to the optional components (extensions) array of the `rust` package. Or at least I think that rustup interprets that?

nrc:
@alexcrichton @brson added to extensions

Arnavion:
I see this is in 2017-02-03's nightly's manifest, but all the `rust-analysis` entries are associated with `target = "x86_64-unknown-netbsd"`

PR #151, opened 2017-02-01
Openssl 1.0.2k

rillian:
Bump openssl, curl, and git to the latest releases.

Openssl has a fix for a medium vulnerability in Diffie-Hellman verification. Otherwise there's nothing critical for security here, it's just good to be up to date.

Be sure to build with `--no-cache` to pick up the Centos 5 update for [CVE-2016-9147](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9147).

### DIFF
--- a/slaves/dist/build_curl.sh
+++ b/slaves/dist/build_curl.sh
@@ -2,8 +2,8 @@
 
 set -ex
 
-VERSION=7.51.0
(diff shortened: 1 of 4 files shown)


rillian:
@alexcrichton How wedded are you to keeping gcc early in the build order? I had to lower security to work around a tls-incompatibility between the GNU mirrors and the stock curl.

PR #152, opened 2017-02-01
Fix TLS incompatibility by building gcc after openssl

rillian:
Follow-up to my build-order question in #151.

Previously we keep gcc at the head of the chain to reduce cycle times working on the container. Asking @alexcrichton again if we can change that.


alexcrichton:
Sounds good to me!

(note that we'll likely stop using this image soon)

PR #153, opened 2017-02-04
Fixed the targets of analysis manifest entries.

Arnavion:
Before this change, all entries would use the last visited target from the
previous loop, thus would not match their parent target section.

---

Ref https://github.com/rust-lang/rust-buildbot/pull/150#issuecomment-277362183


brson:
Thanks! cc @nrc

PR #156, opened 2017-03-07
Update android ndk to r13b

malbarbo:
This allows rustup and cargo be compiled for android on aarch64 and x86.

The old ndk has a compiler that segfaults em compiling openssl. The compiler with r13b ndk works fine.

### DIFF
--- a/slaves/android/Dockerfile
+++ b/slaves/android/Dockerfile
@@ -12,24 +12,25 @@ RUN pip install buildbot-slave
 RUN groupadd -r rustbuild && useradd -r -g rustbuild rustbuild
 RUN mkdir /buildslave && chown rustbuild:rustbuild /buildslave
 
-# Setup PATH to allow running android tools.
-ENV PATH=$PATH:/android/ndk-arm/bin:/android/ndk-aarch64/bin:/android/ndk-x86:/android/sdk/tools:/android/sdk/platform-tools
-
-# Not sure how to install 64-bit binaries in the sdk?
-ENV ANDROID_EMULATOR_FORCE_32BIT=true
-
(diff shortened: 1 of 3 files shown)


malbarbo:
@alexcrichton Maybe you can take a look at this.

I don't know if there is a policy for api level. I used api 21 for ndk because it is the first to support aarch64, that way we use the same api for all archs.

alexcrichton:
Thanks for the PR @malbarbo! Unfortunately though these containers are all but dead at this point (not producing binaries any more). Do you want to send this PR to rust-lang/rust instead? (where we build our release artifacts now). You can find the images in `src/ci/docker/*/Dockerfile` (they should look familiar after changing these).

FWIW we may want to check in with the Gecko folks as they're the ones that requested r9, but if it segfaults then sounds like using r13 would be welcome :)

malbarbo:
Cargo and rustup still uses these containers in their ci. The intention is to add the android container to the ci of cargo and rustup to produce binaries for android. In the future I will take a look in the container
used by the rust ci (there I will try r13).

alexcrichton:
True yeah. Although neither Cargo nor rustup use the Android version? 

FWIW Cargo will from now on be compiled in the rust-lang/rust repository (as part of the dist builders). We don't have long-term plans for rustup, but my guess is that rustup would benefit from copying this configuration into the rustup repo rather than using the images here (which aren't slated to get any more updates)

malbarbo:
Thanks for your attention @alexcrichton. You are right, I will send a PR to rustup copying this configuration. I am busy until the next month... but after that I will take a look at the build integration of cargo and rust.

